PR #14, opened 2017-02-16
Fix parser to conform to spec 0.1-dev

tnozicka:
Fix parser to conform to spec 0.1-dev.
Add validation for excess fields.
Add validation for required fields.
Switch parser from github.com/ghodss/yaml to gopkg.in/yaml.v2 because of UnmarshalYAML method we are using now. (We can't unmarshal it as JSON because int vs. string issues with custom parsing and more.)


pkg/cmd/convert.go
```diff
@@ -8,7 +8,7 @@ import (
 	"reflect"
 	"strings"
 
-	"github.com/ghodss/yaml"
+	kyaml "github.com/ghodss/yaml" // we intentionally use a different yaml package here for kubernetes objects
```
kadel: Can you please add comment explaining why we use different yaml package for k8s objects?

kadel (approved):

PR #16, opened 2017-02-20
update examples

kadel:
- new single container example hello-nginx.yml
 - two container example - simple wordpress with mariadb without external storage


examples/hello-nginx.yaml
```diff
@@ -0,0 +1,7 @@
+version: 0.1-dev
+services:
+- name: helloworld
+  containers:
+  - image: tomaskral/nonroot-nginx
+    ports:
+    - port: 8080
```
tnozicka: missing `\n` at the end

tnozicka (changes requested):

examples/hello-nginx.yaml
kadel: fixed

tnozicka (approved):
LGMT

PR #18, opened 2017-02-20
Remove hostport from opencompose definitions

tnozicka:
It was a mistake to put it there and does not technically make sense.

### DIFF
--- a/pkg/encoding/v1/encoding.go
+++ b/pkg/encoding/v1/encoding.go
@@ -35,7 +35,6 @@ func (rn *ResourceName) UnmarshalYAML(unmarshal func(interface{}) error) error {
 
 type PortMapping struct {
 	ContainerPort int
-	HostPort      int
 	ServicePort   int
 }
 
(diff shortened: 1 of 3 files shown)


kadel (approved):

tnozicka:
This PR broke master!

This is exactly the issue I was mentioning to @kadel in person. Because we no longer force CI checks, the up to date guarantee is also lost which was exactly the case why our CI failed to detect it.

The issue is that both PR #17 and #18 successfully passed a health check after merge with current master at that time, but after merging #17 the check was not run again!

@kadel @pradeepto @surajssd  IMHO **everyone** who is merging a PR and knows that there has been some changes in master since, or just to be sure, **shall restart the CI build** in https://travis-ci.org/redhat-developer/opencompose/ manually **before merging**!

The only other option I know of is even worse and forces everyone to keep their PRs up to date with rebasing (by enforcing CI checks in GitHub settings). 

The root cause of the issue is the GitHub can't be configured (to the best of my knowledge) to do the right thing which would be to invalidate any check that has run against a different master commit than there is now. This would force you to always re-trigger the CI build if there has been a change in master.

PR #29, opened 2017-02-22
rename hello-nginx.yaml to hello-nginx.yml

kadel:
we should use yml extension for all YAML files


tnozicka:
I believe in the exact opposite. We should use `yaml`

kadel:
ok, so it looks like that official recomendation is yaml extension, you are right ;-)

kadel:
closing this :-D

tnozicka:
@kadel also https://github.com/redhat-developer/opencompose/blob/master/examples/wordpress/no-storage.yml has wrong extension

kadel:
yep, I just noticed that. Will fix it

Issue #20, opened 2017-02-21
Extend generated service options


tnozicka:
At this point we always generate k8s service with the default type `ClusterIP`. This was only the first attempt at generating k8s services and we need to get better at this and support type `LoadBalancer`. Although HTTP based services can be exposed using L7 Kubernetes and OpenShift objects (Ingress and Routes - which we don't yet support), non-HTTP services need k8s service of type `LoadBalancer` or with `externalIP` specified to be exposed.

This will affect OpenCompose file format, but we are yet to come up with the syntax.

Ref: 
 - https://kubernetes.io/docs/user-guide/services/
 - https://kubernetes.io/docs/api-reference/v1.5/#servicespec-v1

PR #31, opened 2017-02-22
Extend generated service options to cover internal and external types

tnozicka:
Fixes https://github.com/redhat-developer/opencompose/issues/20


tnozicka:
@pradeepto @kadel this is a file format change. It is backwards-compatible - the new field `type` is optional.

https://github.com/tnozicka/opencompose/blob/51b38c48fc8af3bb815e842629b700bfff5c6dd6/pkg/encoding/v1/encoding.go#L103

kadel (approved):
lgtm

PR #33, opened 2017-02-23
Update go get command

cdrage:
Using HTTPS does not work:

▶ go get -u https://github.com/redhat-developer/opencompose
package https:/github.com/redhat-developer/opencompose: "https://" not
allowed in import path

Thus removing https :)


kadel (approved):
yep, my bad :relaxed:

PR #42, opened 2017-03-01
Clean up installation way for linux releases in README

tnozicka:
cc: @cdrage 

context: https://github.com/redhat-developer/opencompose/issues/36#issuecomment-283330934

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,11 @@ You can retrieve binaries for Linux, macOS and Windows on our [GitHub release pa
 ##### Linux
 If you prefer to do it straight from CLI here's a one-liner for you:
 ```bash
-curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJf - -C ${HOME}/bin ./opencompose
+curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJ ./opencompose
+```
+and if you want it installed system-wide:
+```bash
+sudo mv ./opencompose /usr/local/bin/
 ```
 
 ### Example


README.md
```diff
@@ -27,7 +27,11 @@ You can retrieve binaries for Linux, macOS and Windows on our [GitHub release pa
 ##### Linux
 If you prefer to do it straight from CLI here's a one-liner for you:
 ```bash
-curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJf - -C ${HOME}/bin ./opencompose
+curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJ ./opencompose
+```
+and if you want it installed system-wide:
+```bash
+sudo mv ./opencompose /usr/local/bin/
```
cdrage: why `|`, shouldn't it be `&&` incase the first command fails?

cdrage, line 34: Missing +x and specifying the file output (don't want users to override something by accident)

So:
```
chmod +x ./opencompose
sudo mv ./opencompose /usr/local/bin/opencompose
```

cdrage (changes requested):

README.md
```diff
@@ -27,7 +27,11 @@ You can retrieve binaries for Linux, macOS and Windows on our [GitHub release pa
 ##### Linux
 If you prefer to do it straight from CLI here's a one-liner for you:
 ```bash
-curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJf - -C ${HOME}/bin ./opencompose
+curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJ ./opencompose
```
cdrage, line 30: sohuld be `&&` instead of `|` in case curl fails.

cdrage (changes requested):

README.md
tnozicka: @cdrage It should not!

Then you would have to specify a file to work with to `tar`. The correct way with shell would be:
```bash
set -o pipefail && curl -L https://github.com/redhat-developer/opencompose/releases/download/v0.1.0-alpha.0/opencompose-v0.1.0-alpha.0-d0edfd9-linux-64bit.tar.xz | tar -xJ ./opencompose
```
but I am not putting that in installation instructions. The reason is that this command is meant for interactive use and in case `curl` fails user will see an error from both `curl` and `tar` and it will also give non-zero return code even without the `set -o pipefail` option because if `curl` won't download an archive, `tar` will fail because it has corrupted or missing input.

README.md
tnozicka: @cdrage not missing `+x`. The file is from tar and it already has all the permissions set up correctly.

> specifying the file output (don't want users to override something by accident)

I can't really see how would they do that. Would you mind giving an example?
(For me it's shorter and nicer the way it is but if there is a way to get it easily wrong I can change it.)

README.md
cdrage: Ah I'm an idiot, my early-morning-brain thought it was a ||. Proceed! :)

README.md
cdrage: Not to sure how to give an example, just following usual conventions when copying files over. I've done the mistake personally of auto-tabbing a file when supply `/` and overriding something by accident.

cdrage:
LGTM

cdrage (approved):

kadel (approved):

PR #46, opened 2017-03-03
Add Community section to README

tnozicka:
Now we have ML and Slack (with IRC Bridge).

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,3 +90,6 @@ To make it permanent add this line to your `~/.zshrc`.
 ### Documentation
  - [OpenCompose file reference documentation](https://github.com/redhat-developer/opencompose/blob/master/docs/file-reference.md)
  - You can find more in [doc/](https://github.com/redhat-developer/opencompose/tree/master/docs) folder
+
+### Community
+We always welcome your feedback and thoughts on the project! Come and join our mailing list - [opencompose [at] googlegroups.com](https://groups.google.com/forum/#!forum/opencompose). We also hangout on [slack.k8s.io](http://slack.k8s.io/) ([#sig-apps](https://kubernetes.slack.com/messages/sig-apps/)).


README.md
```diff
@@ -90,3 +90,15 @@ To make it permanent add this line to your `~/.zshrc`.
 ### Documentation
  - [OpenCompose file reference documentation](https://github.com/redhat-developer/opencompose/blob/master/docs/file-reference.md)
  - You can find more in [doc/](https://github.com/redhat-developer/opencompose/tree/master/docs) folder
+
+### Contact
+We always welcome your feedback and thoughts on the project! Come and join us on [Slack](https://opencompose.slack.com) and join our [mailing list](https://groups.google.com/forum/#!forum/opencompose).
+
+#### Slack
+[opencompose.slack.com](https://opencompose.slack.com)
```
pradeepto: Let's not do this please. We should talk on sig-apps on Kubernetes slack instead. If and when time comes, we should get our own channel on the same slack.

pradeepto (changes requested):

README.md
tnozicka: I think we should have both. There should be a channel for discussing the tool, helping users, ... And we should also talk on #sig-apps when we want to engage Kubernetes community; on topics that relate to them. I don't think that spamming #sig-apps with discussing bugs, technological discussions and others would do any good.

kadel: I'm not sure about slack either :-(

1) It is not working as you described ;-) Users need invitation first to sign up to opencompose slack.

2) Isn't ML enough for now?

tnozicka: Ok, Slack sucks. I'll leave only ML and #sig-apps for now.

tnozicka:
@pradeepto @kadel fixed

pradeepto (approved):

pradeepto:
LGTM

PR #100, opened 2017-04-12
Minor change in the description of a test case

concaf:
Minor change in the description of a test case, changed it so
it includes only mandatory fields and removed the mention of
replicas from the name string.


surajssd (approved):
LGTM

PR #105, opened 2017-04-17
add secrets proposal

concaf:
@pradeepto @surajssd @kadel PTAL!

### DIFF
--- a/docs/proposals/secrets-proposal.md
+++ b/docs/proposals/secrets-proposal.md
@@ -42,10 +42,8 @@ services:
 
 secrets:
     ci_secret:
-      travis:
-        file: /etc/travis.passwd
-      jenkins:
-        data: strongpassword@@
+    - file: /tmp/travis.password
+    - data: jenkins=strongpassword@@
 ```
 
 The syntax will look like -
@@ -53,17 +51,13 @@ The syntax will look like -
 ```yaml
 secrets:
   secret_name:
-    secret_key:
-      data: secret_value
-```
+  - file: <file path>
+  - data: <data key>=<data value>
+  - data: <data key>=<base64 encoded data>
+    type: base64
+  - data: <data key>=<data value>
+    type: plaintext (default)
 
-In order to retrieve a secret from a file, use -
-
-```yaml
-secrets:
-  secret_name:
-    secret_key:
-      file: <file path>
 ```
 
 ---
@@ -126,8 +120,7 @@ For the above mentioned syntax, the `secrets` definition at the container level 
 
 `secrets.<secret name>` is a _mandatory_ field of type _string_.
 
-`secrets.<secret name>.<secret key>` is a _mandatory_ field of type _string_. This can either by followed by the `file` or `data` directives, or be provided a _string_ containing the value of the secret key.
-
-`secrets.<secret name>.<secret key>.file` is an _optional_ field of type _string_.
-
-`secrets.<secret name>.<secret key>.data` is an _optional_ field of type _string_.
+`secrets.<secret name>.[]type` is a _mandatory_ filed with allowed keys `file` and `data`.
+`file` contains the path to the file containing secret data
+`data` contains plain text secret data by default along with a plain text key for the value, but a base64 encoded input can also be provided by providing a `type: base64` key with the data.
+All, `file`,`type`, `data` accept `string` values.
--- a/docs/proposals/secrets-proposal.md
+++ b/docs/proposals/secrets-proposal.md
@@ -131,6 +131,6 @@ For the above mentioned syntax, the `secrets` definition at the container level 
 
 `secrets.data.file` is an _optional_ field of type _string_ which takes the path of the file that contain the secret data.
 
-`secrets.base64` is an _optional_ field which takes a _string_ which is base64 encoded as secret data.
+`secrets.data.base64` is an _optional_ field which takes a _string_ which is base64 encoded as secret data.
 
-`secrets.plaintext` is an _optional_ field which takes a _string_ which contains the secret data in plaintext.
+`secrets.data.plaintext` is an _optional_ field which takes a _string_ which contains the secret data in plaintext.
--- a/docs/proposals/secrets-proposal.md
+++ b/docs/proposals/secrets-proposal.md
@@ -27,41 +27,41 @@ To create a new secret, a top level `secrets` key can be used as follows:
 version: "0.1-dev"
 
 services:
-   name: foobar
+ - name: foobar
    containers:
    - image: foo/bar:tag
      ports:
      - port: 8080:80
      env:
-     - foo=bar
-     - travis_creds:
-         secret: ci_secret/travis
+     - name: foo
+       value: bar
+     - name: travis_creds
+       secretRef: ci_secret/travis
      mounts:
      - mountPath: /etc/jenkins.conf
-       secret: ci_secret/jenkins
+       secretRef: ci_secret/jenkins
 
 secrets:
-    ci_secret:
-    - file: /tmp/travis.password
-    - data: jenkins=strongpassword@@
+  - name: ci_secret
+    fromFile:
+    - /tmp/travis.password
+    fromLiteral:
+    - jenkins=strongpassword@@
 ```
 
 The syntax will look like -
 
 ```yaml
 secrets:
-  secret_name:
-  - file: <file path>
-  - data: <data key>=<data value>
-  - data: <data key>=<base64 encoded data>
-    type: base64
-  - data: <data key>=<data value>
-    type: plaintext (default)
-
+  - name: <secret name>
+    fromFile:
+    - <file path>
+    - <file path>
+    fromLiteral:
+    - <data key>=<base64 encoded data>
+    - <data key>=<base64 encoded data>
 ```
-
 ---
-
 
 To consume the above secret in OpenCompose, here is the proposed syntax.
 
@@ -79,11 +79,12 @@ services:
      ports:
      - port: 8080:80
      env:
-     - foo=bar
-     - db_user:
-         secret: kubesec/username
-     - db_pass:
-         secret: kubesec/password
+     - name: foo
+       value: bar
+     - name: db_user
+       secretRef: kubesec/username
+     - name: db_pass
+       secretRef: kubesec/password
 ```
 
 - ##### Consuming an already created Kubernetes secret in OpenCompose as a mounted volume
@@ -99,9 +100,9 @@ services:
      - port: 8080:80
      mounts:
      - mountPath: /var/secret_path/user_info
-       secret: kubesec/username
+       secretRef: kubesec/username
      - mountPath: /var/secret_path/pass_info
-       secret: kubesec/password
+       secretRef: kubesec/password
 ```
 
 ---
@@ -110,17 +111,16 @@ For the above mentioned syntax, the `secrets` definition at the container level 
 
 - environment variables
 
-`containers[].env[].secret` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
+`containers.env.secretRef` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
 
 - volume mounts
 
-`containers[].mounts[].secret` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
+`containers.mounts.secretRef` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
 
 - top level secrets definition
 
-`secrets.<secret name>` is a _mandatory_ field of type _string_.
+`secrets.name` is a _mandatory_ field of type _string_, which sets the name of the secret.
 
-`secrets.<secret name>.[]type` is a _mandatory_ filed with allowed keys `file` and `data`.
-`file` contains the path to the file containing secret data
-`data` contains plain text secret data by default along with a plain text key for the value, but a base64 encoded input can also be provided by providing a `type: base64` key with the data.
-All, `file`,`type`, `data` accept `string` values.
+`secrets.fromFile` is an _optional_ field which takes an _array_ of file paths that contain the secret data.
+
+`secrets.fromLiteral` is an _optional_ field which takes an _array_ of strings, of the form _plain text key=base64 encoded secret data_
--- a/docs/proposals/secrets-proposal.md
+++ b/docs/proposals/secrets-proposal.md
@@ -1,0 +1,140 @@
+## Proposal for supporting secrets in OpenCompose
+
+### Kubernetes Secrets
+
+A very simple secret object in Kubernetes looks like this -
+
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: kubesec
+type: Opaque
+data:
+  username: <base64 encoded value>
+  password: <base64 encoded value>
+```
+
+Note that, as seen above, one secret can have multiple `data` values.
+
+---
+
+To consume the above secret in OpenCompose, here is the proposed syntax.
+
+Secrets in Kubernetes can be consumed either as an environment variable, or as a mounted volume.
+ 
+- ##### Consuming an already created Kubernetes secret in OpenCompose as an environment variable
+ 
+ ```yaml
+version: "0.1-dev"
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     env:
+     - foo=bar
+     - db_user:
+         secret: kubesec/username
+     - db_pass:
+         secret: kubesec/password
+```
+
+- ##### Consuming an already created Kubernetes secret in OpenCompose as a mounted volume
+ 
+ ```yaml
+version: "0.1-dev"
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     mounts:
+     - mountPath: /var/secret_path/user_info
+       secret: kubesec/username
+     - mountPath: /var/secret_path/pass_info
+       secret: kubesec/password
+```
+
+---
+
+#### Creating a new secret:
+
+To create a new secret, a top level `secrets` key can be used as follows:
+
+ ```yaml
+version: "0.1-dev"
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     env:
+     - foo=bar
+     - travis_creds:
+         secret: ci_secret/travis
+     mounts:
+     - mountPath: /etc/jenkins.conf
+       secret: ci_secret/jenkins
+
+secrets:
+    ci_secret:
+      travis:
+        file: /etc/travis.passwd
+      jenkins:
+        value: strongpassword@@
+      circle: stillstrongpassword@@
+```
+
+Here, the short syntax treats the provided value as a secret literal.
+```yaml
+secrets:
+  secret_name:
+    secret_key: secret_value
+```
+
+Or, the long syntax will be more verbose like -
+
+```yaml
+secrets:
+  secret_name:
+    secret_key:
+      value: secret_value
+```
+
+In order to retrieve a secret from a file, use -
+
+```yaml
+secrets:
+  secret_name:
+    secret_key:
+      file: <file path>
+```
+
+---
+
+For the above mentioned syntax, the `secrets` definition at the container level will have the following fields -
+
+- environment variables
+
+`containers[].env[].secret` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
+
+- volume mounts
+
+`containers[].mounts[].secret` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
+
+- top level secrets definition
+
+`secrets.<secret name>` is a _mandatory_ field of type _string_.
+
+`secrets.<secret name>.<secret key>` is a _mandatory_ field of type _string_. This can either by followed by the `file` or `value` directives, or be provided a _string_ containing the value of the secret key.
+
+`secrets.<secret name>.<secret key>.file` is an _optional_ field of type _string_.
+
+`secrets.<secret name>.<secret key>.value` is an _optional_ field of type _string_.
--- a/docs/proposals/secrets-proposal.md
+++ b/docs/proposals/secrets-proposal.md
@@ -43,10 +43,13 @@ services:
 
 secrets:
   - name: ci_secret
-    fromFile:
-    - /tmp/travis.password
-    fromLiteral:
-    - jenkins=strongpassword@@
+    data:
+    - key: travis
+      file: /tmp/travis.password
+    - key: jenkins
+      plaintext: strongpassword@@
+    - key: db
+      base64: YWRtaW4=
 ```
 
 The syntax will look like -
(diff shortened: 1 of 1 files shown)


docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+```
+
+Or, the long syntax will be more verbose like -
+
+```yaml
+secrets:
+  secret_name:
+    secret_key:
+      value: secret_value
+```
```
pradeepto: @containscafeine What is the advantage of having both long and short forms? Can we not just have one? That was one of the tenets of OpenCompose - to be simple, to have one way to doing things. What do you think? 

Having said that, if there is a very important reason, I would be ok with it.

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,140 @@
...
+       secret: ci_secret/jenkins
+
+secrets:
+    ci_secret:
+      travis:
+        file: /etc/travis.passwd
+      jenkins:
+        value: strongpassword@@
+      circle: stillstrongpassword@@
+```
```
pradeepto: Can you move this example up? I mean, it makes more sense to have this example (creation of secret) before the one that shows how to consume a secret.

pradeepto (commented):
Also, would you like to cover imagePullSecrets or should that be covered somewhere else?

docs/proposals/secrets-proposal.md
concaf: fixed

docs/proposals/secrets-proposal.md
concaf: @pradeepto the short syntax sort of makes writing the spec more intuitive, and assumes intelligent defaults. The long syntax provides more knobs to play with, but is used rarely.

One good example would be the `build:` directive in Docker Compose - https://docs.docker.com/compose/compose-file/#build

- short syntax -

```yaml
build: ./dir
```
- long syntax -

``` yaml
build:
  context: ./dir
  dockerfile: Dockerfile-alternate
  args:
    buildno: 1
```

Personally, this sort of definition is very intuitive and clean to write, WDYT?

pradeepto: I understand the intent but remember two ways of doing things introduces complexity. As I said, I am not against it if there is a good rationale for it and others agree.

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,141 @@
...
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     env:
+     - foo=bar
+     - db_user:
+         secret: kubesec/username
+     - db_pass:
+         secret: kubesec/password
```
surajssd: what gets exposed inside the container?

surajssd:
So in multiple places we have two ways of doing things?

surajssd:
Needs a better way to mention `secrets`(root level `secrets`) in the opencompose file, because if someone is writing the `secrets` in plain text why would somebody not use the `env` directly?

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,141 @@
...
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     env:
+     - foo=bar
+     - travis_creds:
```
surajssd: When using the yaml keys as user defined the problem is it's hard to know what should be user defined and what is from the spec?

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,141 @@
...
+         secret: ci_secret/travis
```
surajssd: Also when we have time let's not do string magic?

surajssd:
Right now kubernetes does not have any field where it asks users to state yaml keys, in all places it provides concrete keys:

Asking user to provide keys means:
```yaml
services:
  foo:
  ...
  bar:
  ...
```

Opencompose providing concrete keys:
```yaml
services:
- name: foo
  ...
- name: bar
  ...
```

What are the advantages or disadvantages of both on user side on our side as creator of this spec and in terms of expanding this format?

So this argument because with this proposal the `envs` will be handled like that!

surajssd:
>So in multiple places we have two ways of doing things?

so in terms of implementation how good or bad about this? Because `libcompose` did lot of hacks and workarounds to implement it.

surajssd:
What are the advantages or disadvantages of fetching content of file and giving only path to file in opencompose?

Like it's done in this proposal right now in root level `secrets` where key is `file` ?

kadel:
> Needs a better way to mention secrets(root level secrets) in the opencompose file, because if someone is writing the secrets in plain text why would somebody not use the env directly?

One of the reason is that one secret can be shared between multiple services. For example db password between multiple services accessing the same database.

docs/proposals/secrets-proposal.md
kadel: Agreed. keys shouldn't hold users data. They should describe what kind of data is stored in value.

kadel:
> Right now kubernetes does not have any field where it asks users to state yaml keys, in all places it provides concrete keys:
> 
> Asking user to provide keys means:
> 
> services:
>   foo:
>   ...
>   bar:
>   ...
> Opencompose providing concrete keys:
> 
> services:
> - name: foo
>   ...
> - name: bar
>   ...
> What are the advantages or disadvantages of both on user side on our side as creator of this spec and in terms of expanding this format?
> 
> So this argument because with this proposal the envs will be handled like that!

We  are  using  this form
```
services:
 - name: foo
```
Because this is how its used in most places. you always have `key: value` pair. Where `key` is predefined by file spec, and a user  just sets `value`.
with 

```
services:
  foo:
```

the user is setting data in `key` but inside that key, you are using again `key: value` with predefined `key`.

This means that you are combining two ways how to write things.

I think we should avoid that as we done so far.Sometimes it means to make things a bit longer, but that's OK as log as everything is consistent and readable.

Kubernetes follows the same convention, except one place and that is `data` in Secrets.
```
data:
  username: <base64 encoded value>
  password: <base64 encoded value>
```
haven't found a reason why they did it here like this (user input in key), but in `env` they are doing
```
   env:
    - name: foo
      value: bar
```

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,133 @@
...
+      data: secret_value
+```
+
+In order to retrieve a secret from a file, use -
+
+```yaml
+secrets:
+  secret_name:
+    secret_key:
+      file: <file path>
```
kadel: This is not something that is not in Kubernetes Secret object. 
Is this new feature that we are going to support in OpenCompose?

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,133 @@
...
+     env:
+     - foo=bar
+     - travis_creds:
+         secret: ci_secret/travis
+     mounts:
+     - mountPath: /etc/jenkins.conf
+       secret: ci_secret/jenkins
+
+secrets:
+    ci_secret:
```
kadel: as discussed in other comments here, keys should have a predefined name and shouldn't store data that users inputs.

I would do it like this so  its consistent with other top level sections like `services` and `volumes`:
```
secrets:
  - name: ci_secret
    data:
       - name: jenkins
         value: stringpasswrd
       - name: travis
         file: /etc/travis.passwd
       - name: password
         valueBase64: YXNmZA0KYXNkZg==
```

This is consistent with other parts of OpenCompose spec.
`value`, `file` and `valueBase64` can't be used together in one data field.
`valueBase64` is there so the user can create secrets that contain non-string data (for example binary).

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,133 @@
...
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     env:
+     - foo=bar
+     - travis_creds:
```
kadel: This is kind of strange, as now we have two different types (dict and string) inside env array :-(
Maybe we should reconsider whole `env` structure.

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,133 @@
...
+
+services:
+   name: foobar
+   containers:
+   - image: foo/bar:tag
+     ports:
+     - port: 8080:80
+     mounts:
+     - mountPath: /var/secret_path/user_info
+       secret: kubesec/username
```
kadel: `secretName` might be better as it will be consistent with how we reference Volumes here (`volumeName`).

docs/proposals/secrets-proposal.md
concaf: @surajssd the mentioned data key of the given secret. The format is `secret_name/data_name`

docs/proposals/secrets-proposal.md
concaf: IMO this is intuitive. If we keep on adding YAML keys for every data input, I'm not sure if that will make writing the spec any easier.

docs/proposals/secrets-proposal.md
concaf: In that case, we add as many keys are required in Kubernetes configuration itself. How are we making this easier and more intuitive?

concaf:
@surajssd - 

> So in multiple places we have two ways of doing things?

Which 2 ways are you referring to?

> Needs a better way to mention secrets(root level secrets) in the opencompose file, because if someone is writing the secrets in plain text why would somebody not use the env directly?

We can enforce the secret data to be base64 encoded, which is at least a bit better than plaintext. However, users can specify file names if they don't want to put the secret data in the OpenCompose file itself. WDYT?

> What are the advantages or disadvantages of both on user side on our side as creator of this spec and in terms of expanding this format?

Well, every concrete spec defined key means the end user has to learn a new key. If we just make it simpler and more intuitive, I don't see why user defined keys is a problem.

> so in terms of implementation how good or bad about this? Because libcompose did lot of hacks and workarounds to implement it.

The implementation diverts from the regular flow, yes, but if it makes it simpler for the user, that's what matters, no?

> What are the advantages or disadvantages of fetching content of file and giving only path to file in opencompose?
Like it's done in this proposal right now in root level secrets where key is file ?

Well, the user does not have to write the secret in the spec file and can commit it to a VCS if he specifies a file path.

concaf:
@kadel -

> Because this is how its used in most places. you always have key: value pair. Where key is predefined by file spec, and a user just sets value.
...
This means that you are combining two ways how to write things.
I think we should avoid that as we done so far.Sometimes it means to make things a bit longer, but that's OK as log as everything is consistent and readable.

I agree, till now all the keys are set by the spec, and values from the user. But does that make it simpler for the user?
Yes, it defining all the keys makes the things a lot longer, but imo either ways are readable enough.

> Kubernetes follows the same convention

Yep, but should we follow the same? Kubernetes' definitions are not the most intuitive; tbh I'd rather look at Docker Compose spec for the simplicity of defining an application. It's readable as well intuitive. WDYT?

docs/proposals/secrets-proposal.md
concaf: Kubernetes does not support it from the secret spec, but it allows passing in secrets from the command line - `kubectl create secret generic kadel --from-file=/var/lib/super.secret`, so I have added this in the spec. Also, this makes it possible to commit the OpenCompose file to a VCS.
WDYT?

P.S. - I have made some changes in  this part of the spec, I will be updating the proposal.

docs/proposals/secrets-proposal.md
concaf: +1 for adding in base64 field

docs/proposals/secrets-proposal.md
concaf: @kadel - yep, I implemented this by unmarshalling to `interface{}` and then doing the type checking.
I don't think we should change `key=value` because that is the de facto for passing in env variables in most specs, and we will need something else for defining an env secret.

docs/proposals/secrets-proposal.md
concaf: yep, need to add volumeName here

concaf:
TODO :

- [x] update proposal for new root level secrets definition
- [x] add passing base64 value for secrets

docs/proposals/secrets-proposal.md
kadel: We are making it easier and more intuitive by making it consistent.

It's confusing if different parts of OpenCompose specs follow different rules.

docs/proposals/secrets-proposal.md
kadel: To be honest I think this can't stay as it is. It just creates a lot of confusion. 

I think that the way you are referencing secret here is ok, we will have to change the old `key=value` string and come with something else.

How about doing it similar to  Kubernetes?
We don't have to reinvent everything just for the sake of creating something new. Let's stay with k8s way of doing things where it makes sense.

how about this?
```
  env:
     - name: bar
        value: foo
     - name: travis_creds
        secret:  ci_secret/travis 
```

docs/proposals/secrets-proposal.md
kadel: What I meant was change `secret` to `secretName` ;-) `s/secret/secretName`

kadel:
> I agree, till now all the keys are set by the spec, and values from the user. But does that make it simpler for the user?

It makes much more simpler because it's consistent in all part of the spec ;-)

kadel:
> Yep, but should we follow the same? Kubernetes' definitions are not the most intuitive; tbh I'd rather look at Docker Compose spec for the simplicity of defining an application. It's readable as well intuitive. WDYT?


Kubernetes definitions are quite intuitive is if think about that, its just a lot of them and that is what we are trying to solve here. Docker compose can be also confusing in some ways, there are same things that can be written in multiple ways, for example, look at env varibales there.

In my opinion intuitive can be only achieved if the whole spec is consistent, the whole spec should follow same rules.

docs/proposals/secrets-proposal.md
concaf: @kadel I agree that this part is not consistent, but env variables are set at `key=value` and it's just not very natural defining them as -
```yaml
name: foo
value: bar
```

How about -
```yaml
env:
- foo: bar
- travis_creds:
    secret: ci_secret/travis
```

or

```yaml
env:
- foo: bar
- travis_creds: ci_secret/travis
  type: secret
```

concaf:
@kadel -
> Kubernetes definitions are quite intuitive is if think about that, its just a lot of them and that is what we are trying to solve here

Yep, but defining a key for every little knob in the configuration makes it crazy long, and sort of defeats the purpose of a simpler, concise and readable spec :(

> Docker compose can be also confusing in some ways, there are same things that can be written in multiple ways, for example, look at env varibales there

Agree that we should not provide multiple ways of doing the same thing, but I don't think it is confusing.

> In my opinion intuitive can be only achieved if the whole spec is consistent, the whole spec should follow same rules.

Yep, totally agree. I'd probably want the rest of the spec to be more user friendly and consistent at the same time, and now is a good time to change it.

concaf:
- [x] Do not error out if root level secrets are not present, instead log that we are assuming that the mentioned secrets are present in the cluster.
- [x] Make `services.containers[].mounts.volumeRef` optional if `services.containers[].mounts.secretRef` is present
- [x] Should we remove support for plain text values in the root level secrets definition? - we decided, no!

concaf:
So after the recent discussions, we have standardized on taking control of the keys in the spec, and I have updated the PR according to that.
@surajssd @kadel PTAL!

docs/proposals/secrets-proposal.md
```diff
@@ -27,42 +27,42 @@ To create a new secret, a top level `secrets` key can be used as follows:
...
-  - data: <data key>=<base64 encoded data>
-    type: base64
-  - data: <data key>=<data value>
-    type: plaintext (default)
-
+  - name: <secret name>
+    fromFile:
+    - <file path>
+    - <file path>
+    fromLiteral:
```
kadel: This key name sound strange, at least to me :-(
Isn't there a better name?

Isn't `data` enough?

concaf: @kadel agree, I took these from from the `kubectl` command line flags for secrets, which are `--from-file` and `--from-literal`.

`data` is fine but it might confuse with the `secrets.data` field in Kubernetes secrets. But yes, we could go ahead with `data` since it's a generic enough term.

Also, if we go ahead with `data`, then I'd also want the `from` prefix to be removed from the file so the fields look the same a bit, i.e. `data` and `file` instead of `data` and `fromFile`.

WDYT?

kadel: > data is fine but it might confuse with the secrets.data field in Kubernetes secrets. But yes, we could go ahead with data since it's a generic enough term.

Kubernetes `secrets.data` field is doing exactly the same thing as our `secrets.data` no? 


> Also, if we go ahead with data, then I'd also want the from prefix to be removed from the file so the fields look the same a bit, i.e. data and file instead of data and fromFile.


I have one issue with file/fromFile. And that it's not obvious that the name will be taken from filename :-(
This can be confusing :-(

kadel: what if we do it a bit differently?

What do you think about this?

```yaml
secrets:
  - name: <secret_name>
    data:
     - name: <data key>
       value: <data value>
     - name: <data key>
       file: <file path>
     - name: <data key>
       valueBase64: <base64 value>
```

concaf: @kadel -
> I have one issue with file/fromFile. And that it's not obvious that the name will be taken from filename :-(

Agree, that is somehow a convention that Kubernetes uses, but I can see it's not immediately obvious.

How about -
```yaml
secrets:
  - name: <secret name>
    data:
    - key: <secret data key>
      value: <base64 encoded secret data value>
    - key: <secret data key>
      file: <path to file containing secret data>
```

concaf:
@kadel @surajssd also, does it make sense to support taking in plain text secret data values? Should we only accept base64 encoded values as input, or both?

docs/proposals/secrets-proposal.md
kadel: Yep, structuring it like this might be best.

I was just thinking that it would be nice to give users option where they don't have to do base64 encoding for every value.
I expect that most user will use some basic string as value, hence two different keys. `value` for plain string and `valueBase64` for advanced use-cases (binary data, multiline etc..) where it's required to have  base64 encoded value.

This was just my rationale behind having `value` and `valueBase64`.
I'm ok with keeping it as `value` that is always base64 encoded.

pradeepto: Couple of things: 
1) Whatever we do, let's remember the "one way to do things" guideline before we implement anything.
2) But if (ignoring the above guideline) we add both - plaintext and base64 encoded, then let's make sure the tooling WARNS the user that "bro, this is not cool".

concaf: @kadel -
Agree with you, imo, let's do both. I'll proceed with something like this -
```yaml
secrets:
  - name: <secret name>
    data:
    - key: <secret data key>
      base64: <base64 encoded secret data value>
    - key: <secret data key>
      file: <path to file containing secret data>
    - key: <secret data key>
      plaintext: <secret data value in plaintext>

```

kadel: > But if (ignoring the above guideline) we add both - plaintext and base64 encoded, then let's make sure the tooling WARNS the user that "bro, this is not cool".

only one of them can be set at the time

kadel: @containscafeine 

what is wrong with `value` and `valueBase64`?

 `plaintext` and `base64`  looks a bit wage to me. It is not saying what the `plaintext` is. It should describe how this field is related to `key` field. 

In the same sense, I don't even like using `file` but I couldn't think of anything better. Maybe  `valueFromFile` , but using 3 words for it seems too much.

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+
+`secrets.name` is a _mandatory_ field of type _string_, which sets the name of the secret.
+
+`secrets.data` is a _mandatory_ field, which contains the following information about the secret.
+
+`secrets.data.key` is a _mandatory_ field of type _string_, which sets the the secret data key in the secret.
+
+`secrets.data.file` is an _optional_ field of type _string_ which takes the path of the file that contain the secret data.
+
+`secrets.base64` is an _optional_ field which takes a _string_ which is base64 encoded as secret data.
```
surajssd: `s/secrets.base64/secrets.data.base64`

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+
+`secrets.plaintext` is an _optional_ field which takes a _string_ which contains the secret data in plaintext.
```
surajssd: `s/secrets.plaintext/secrets.data.plaintext`

surajssd:
Also mention about the fields under `data` being optional but either one of them have to be specified? And what happens when multiple of them are given or is that even allowed?

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+
+For the above mentioned syntax, the `secrets` definition at the container level will have the following fields -
+
+- environment variables
+
+`containers.env.secretRef` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
+
+- volume mounts
+
+`containers.mounts.secretRef` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
```
surajssd, line 122: While implementing you will have to also change the `volumeRef` to be optional which is not right now.

Also make sure that you mention this is optional field but either `volumeRef` or `secretRef` should be present.

docs/proposals/secrets-proposal.md
```diff
@@ -0,0 +1,136 @@
...
+       secretRef: kubesec/password
+```
+
+---
+
+For the above mentioned syntax, the `secrets` definition at the container level will have the following fields -
+
+- environment variables
+
+`containers.env.secretRef` is an _optional_ field of type _string_. The string will consist of 2 parts separated by a `/`, the first being the secret name, and the second part being the key in the secret object.
```
surajssd, line 118: Like I mentioned in the [comment below](https://github.com/redhat-developer/opencompose/pull/105/files#r117732117), here also mention that this is optional field, but either `secretRef` or `value` should be present.

surajssd:
Creating a topic of conflicting field like this might help.

https://github.com/surajssd/opencompose/blob/cd34147a553fe8b46699e0d66dd134e166027ae9/docs/proposals/healthchecks.md#conflicting-fields

screenshot from 2017-05-22 18-02-59

docs/proposals/secrets-proposal.md
kadel: ping @containscafeine ^

kadel:
> @kadel @surajssd also, does it make sense to support taking in plain text secret data values? Should we only accept base64 encoded values as input, or both?

I was thinking about allowing plaintext values as a convenience for users.
Base64 doesn't add any security, it's there just so you can set values with something that is not just plaintext. 
But if you think that we don't need plaintext we can do just base64 encoded values

docs/proposals/secrets-proposal.md
surajssd: ping @containscafeine

docs/proposals/secrets-proposal.md
surajssd: ping @containscafeine

concaf: fixed

docs/proposals/secrets-proposal.md
concaf: fixed

surajssd:
It would be great if we consider adding `envFrom`, this saves a lot of yaml writing.

Here you just refer to a secret and then all the envs from that secret get imported into the current container.

```bash
$ kubectl explain deployment.spec.template.spec.containers.envFrom
RESOURCE: envFrom <[]Object>

DESCRIPTION:
     List of sources to populate environment variables in the container. The
     keys defined within a source must be a C_IDENTIFIER. All invalid keys will
     be reported as an event when the container is starting. When a key exists in
     multiple sources, the value associated with the last source will take
     precedence. Values defined by an Env with a duplicate key will take
     precedence. Cannot be updated.

    EnvFromSource represents the source of a set of ConfigMaps

FIELDS:
   secretRef    <Object>
     The Secret to select from

   configMapRef <Object>
     The ConfigMap to select from

   prefix       <string>
     An optional identifer to prepend to each key in the ConfigMap. Must be a
     C_IDENTIFIER.
```

kadel:
> It would be great if we consider adding envFrom, this saves a lot of yaml writing.

I don't think this should be in this initial implementation.
Let's keep this in mind for future improvements.

docs/proposals/secrets-proposal.md
concaf: took care of this during implementation #113

docs/proposals/secrets-proposal.md
concaf: #113 take cares of this

PR #113, opened 2017-05-05
Implement secrets

concaf:
Up for review


pkg/goutil/goutil.go
```diff
@@ -19,3 +19,12 @@ func Int32Addr(i int32) *int32 {
 func BoolAddr(b bool) *bool {
 	return &b
 }
+
+func IsElementInArray(e interface{}, a []interface{}) bool {
```
kadel: It looks like this function is not used anywhere.

pkg/object/object.go
```diff
@@ -70,10 +75,30 @@ type Volume struct {
...
+}
+
 type OpenCompose struct {
 	Version  string
 	Services []Service
 	Volumes  []Volume
+	Secrets  []Secret
+}
+
+type Input struct {
```
kadel: What is this? This shouldn't be here.

glide.lock
```diff
@@ -169,4 +184,63 @@ imports:
   - pkg/util/yaml
   - pkg/watch
   - pkg/watch/versioned
+- name: k8s.io/kubernetes
```
kadel: :-(  this is not good, it's adding a looot of unnecessary dependencies :(
just binary size increases from 35MB to 50MB :-(

pkg/object/object.go
concaf: I have created this and changed the definition of `func (d *Decoder) Decode(in *object.Input) (*object.OpenCompose, error)`. I have done this because in case in the OpenCompose file the relative path of the secret is passed, then we need resolve that before passing to kubernetes.go.

glide.lock
concaf: Yup :(
But this lets us add validations for the generated Kubernetes objects in kubernetes.go.
Currently, I am only doing this for secrets, but using this package we can validate all the generated artifacts.
This is referenced in #139, and this is how I am doing this for secrets -
```golang
	validationErrors := kvalidation.ValidateSecret(&kapi.Secret{
		ObjectMeta: kapi.ObjectMeta{
			Name: sec.ObjectMeta.Name,
			// Putting namespace as default here because ValidateSecret errors
			// out if no namespace is provided
			Namespace: "default",
		},
		Data: sec.Data,
	})
```

kadel: It might be better to do this in another PR after we get secrets in.

pkg/object/object.go
kadel: it definitely need some comments explaining what it is ;-)
Does it have to be part of `object`?

pkg/goutil/goutil.go
concaf: Ack, will remove it.

glide.lock
concaf: @kadel the vendor and glide commit is different with a relevant commit message, and this is commit is after and is relevant to secrets.
We can add similar validations for other objects in subsequent PRs, but since this is relevant to secrets and testing the same, should it not be fine if this remains in this PR in this commit?

kadel: I'm still not sure if it is worth it to add it. Especially at the cost of adding half of the kubernetes as a dependency. I would like to avoid that if possible.
There is a big potential for it to cause big mess in our dependencies in combination with `client-go` and with OpenShift in the future.

concaf: @kadel I agree on the size part, but I'd really want this to happen, since this adds so much value and itegrity to the generated artifacts, in the sense that they will not fail to get created on the cluster.
Is there some other way we can do this? Maybe by not making this a part of the code, or running these only in tests, or some way?

kadel: this is why I would like this to be separate PR/issue so we can discuss this without blocking secrets

pkg/goutil/goutil.go
concaf: Removed

pkg/object/object.go
concaf: Added comments to explain this.
As for it being a part of object.go, I'd ideally want this to be a part of the `cmd` package, but then that messes up the go import cycle when this is used in `encoding.go`, so I made this a part of `object.go`

glide.lock
concaf: tracking this in #139, removed it from the code

pkg/object/object.go
```diff
@@ -227,9 +284,82 @@ func (v *Volume) validate() error {
...
+		count++
+	}
+	if sd.Base64 != nil {
+		count++
+	}
+	if sd.File != nil {
+		count++
+	}
+
+	switch count {
```
kadel, line 313: why `switch` here? Wouldn't be better to use `if`?

If you really want `switch` that there should be  `default`  case to make ti future proof.

pkg/object/object.go
kadel: hmm :-( It really can't be in another place? It would be nice to keep `object` clean and have  there only stuff related to spec.

pkg/object/object.go
concaf: `switch` just seemed a bit idiomatic to me, but meh, just personal preference.
I have added the `default` :+1:

pkg/encoding/v1/encoding.go
```diff
@@ -481,5 +553,47 @@ func (d *Decoder) Decode(data []byte) (*object.OpenCompose, error) {
...
+			// if it is a relative path
+			// and
+			// the OpenCompose input file is either STDIN or file path
+			// then
+			// convert it to an absolute file path.
+			// This will fail if input file is URL
+			if (secData.File != nil) && !filepath.IsAbs(*secData.File) {
+
+				if in.URL != nil {
+					return nil, fmt.Errorf("Unable to decode secret %v with key %v has a relative file path %v, but is being passed as a URL", secret.Name, secData.Key, secData.File)
```
kadel: this message doesn't make sense to me :-(
its show like this:
`ERROR: could not unmarshal data for file 'https://raw.githubusercontent.com/containscafeine/opencompose/14c98c39e2183783ccb356085847a61da5963ed3/examples/wordpress/storage-secrets.yaml': Unable to decode secret dbcreds with key root_password has a relative file path 0xc42043a170, but is being passed as a URL`

If I'm that sees just this message I'm just confused :(

What you want to say here is that it is not possible to load opencompose file from URL that loads secrets from file right? It might be better to say it like this ;-)

pkg/object/object.go
kadel:
```
▶ ./opencompose -f examples/wordpress/storage-secrets.yaml convert -o -
ERROR: failed to validate secret dbcreds: failed to validate secret key root_password: Something went wrong with counting the secret fields for the secret key: root_password
```

you have to also move  `return nil` from  the end to `case 1`:  `case 1: return nil`
You want this to succeed only if count is 1.

pkg/encoding/v1/encoding.go
kadel: Other possible solution would be that we allow loading secrets from a file when opencompose file is loaded from URL.
It will load secrets from a local file, if a path is relative it will be relative to current directory.

pkg/object/object.go
```diff
@@ -113,17 +144,34 @@ func (e *EnvVariable) validate() error {
...
 	return nil
 }
 
 func (m *Mount) validate() error {
 
-	// validate volumeRef
-	if err := validateName(m.VolumeRef); err != nil {
-		return fmt.Errorf("mount %q: invalid name, %v", m.VolumeRef, err)
+	// make sure mountRef and SecretRef are not supplied at the same time
+	// also, that at least one of the both should be specified
```
kadel, line 169: "one of the both"?

pkg/object/object.go
```diff
@@ -236,9 +295,84 @@ func (v *Volume) validate() error {
...
+
+				dataKeyFound := false
+				for _, secData := range secret.Data {
+					if secretDef.DataKey == secData.Key {
+						dataKeyFound = true
+						break
+					}
+				}
+				if !dataKeyFound {
+					log.Printf("Root level secret name : %v found, but the provided data key : %v is missing, assuming the corresponding secret exists in the cluster", secret.Name, secretDef.DataKey)
```
kadel: please remove space before ':'

pkg/transform/kubernetes/kubernetes.go
```diff
@@ -217,23 +243,53 @@ func (t *Transformer) CreateDeployments(s *object.Service) ([]runtime.Object, er
...
 				SubPath:   mount.VolumeSubPath,
 			}
 
+			if mount.VolumeRef != nil {
+				volumeMount.Name = *mount.VolumeRef
+			}
+
+			if mount.SecretRef != nil {
+				volumeMount.ReadOnly = true
+
```
kadel:
```yaml
        volumeMounts:
        - mountPath: /tmp/asdf
          name: ""
          readOnly: true
```
invalid deployment gets with volumeMoutns that look like this generated because you are not setting `volumeMount.name` here.

```go
volumeMount.Name = secretDef.SecretName
```

pkg/encoding/v1/encoding.go
concaf: Ack, I will reword the error.

> It will load secrets from a local file, if a path is relative it will be relative to current directory.

I'm not sure if this is a good idea, since most use cases for this will be users putting in URL from their VCS repositories, and the secret file will most likely be present in the VCS.
Wait, but yeah, it would make sense because the secret file will not be committed to the VCS and will be in .gitignore, but might be present locally.

Yep, I think, we should log this message that we will be using the secret from the local path, and use it.

Yep

kadel: We can start with just rewording error message. Loading files from local filesystem can be done later.

pkg/object/object.go
concaf: forgot to run the tests, fixed

pkg/encoding/v1/encoding.go
concaf: fixed, merged this with `in.STDIN` condition and logged the message

pkg/object/object.go
concaf: fixed, reworded

pkg/object/object.go
concaf: fixed

pkg/transform/kubernetes/kubernetes.go
concaf: Great catch :)
I am surprised that no test caught this, I have added a test for this.

concaf:
- [x] Need to add docs

pkg/transform/util/util.go
```diff
@@ -0,0 +1,24 @@
...
+	"fmt"
+	"strings"
+)
+
+type SecretDef struct {
+	SecretName string
+	DataKey    string
+}
+
+func SecretRefToSecretDef(secretRef *string) (*SecretDef, error) {
```
surajssd: I am not sure about keeping this here, because `util` generally should be something that is generic and can be used in varied number of places.

pkg/object/object.go
```diff
@@ -70,10 +75,36 @@ type Volume struct {
...
+// store the data input metadata which is used in other parts of the code
+// to make decisions.
+// For instance, this is being set in cmd.GetValidatedObject(), and getting
+// in v1.Decode() to convert the relative path of the secret file from the
+// OpenCompose file, and convert it to an absolute path
+type Input struct {
+	Data     []byte
+	STDIN    bool
+	URL      *string
+	FilePath *string
```
surajssd, line 112: Any specific reason why URL and FilePath are pointers?

pkg/object/object.go
```diff
@@ -236,9 +295,85 @@ func (v *Volume) validate() error {
 	return nil
 }
 
+func (sd *SecretData) validate() error {
```
surajssd, line 301: Where do you validate if the field `Base64` is a valid base64 encoded value? 

Similarly for field `File`, it should be a valid path, either absolute or relative depending on what is supported!

pkg/encoding/v1/encoding.go
```diff
@@ -481,5 +555,46 @@ func (d *Decoder) Decode(data []byte) (*object.OpenCompose, error) {
...
+		for _, secData := range secret.Data {
+
+			// If the secret has been provided through a file
+			// and
+			// if it is a relative path
+			// and
+			// the OpenCompose input file is either STDIN or file path
+			// then
+			// convert it to an absolute file path.
+			// This will fail if input file is URL
```
surajssd: why are these comments little weird? One word on one line?

pkg/encoding/v1/encoding.go
```diff
@@ -440,15 +501,28 @@ func (d *Decoder) Decode(data []byte) (*object.OpenCompose, error) {
...
-				})
+				env := object.EnvVariable{
+					Key: e.Key,
+				}
+
+				if e.Value != nil {
+					env.Value = e.Value
+				}
+
+				if e.SecretRef != nil {
```
surajssd: this should be `else if` of the above right? because only one field should be valid right at a time? either `Value` or `SecretRef`? And the `else` should be an error?

pkg/encoding/v1/encoding.go
```diff
@@ -481,5 +555,46 @@ func (d *Decoder) Decode(data []byte) (*object.OpenCompose, error) {
...
+
+			// If the secret has been provided through a file
+			// and
+			// if it is a relative path
+			// and
+			// the OpenCompose input file is either STDIN or file path
+			// then
+			// convert it to an absolute file path.
+			// This will fail if input file is URL
+			if (secData.File != nil) && !filepath.IsAbs(*secData.File) {
```
surajssd, line 612: I think I don't understand what is happening here in this `if` block.

pkg/encoding/v1/encoding.go
```diff
@@ -190,14 +194,15 @@ type ImageRef string
...
-	MountPath string       `yaml:"mountPath"`
+	VolumeRef *ResourceName `yaml:"volumeRef,omitempty"`
+	MountPath string        `yaml:"mountPath"`
 	// these are optional fields so making them as pointer because it helps
 	// to identify whether these fields were given by user or not
 	// if these are not pointer then it is hard to identify what was given
 	// by user and what is the default value
 	VolumeSubPath *string `yaml:"volumeSubPath,omitempty"`
 	ReadOnly      *bool   `yaml:"readOnly,omitempty"`
+	SecretRef     *string `yaml:"secretRef,omitempty"`
```
surajssd: can this be of type `ResourceName`, similar to `VolumeRef`?

pkg/encoding/v1/encoding_test.go
```diff
@@ -1005,21 +1138,6 @@ services:
 		{
 			false, `
 version: 0.1-dev
-services:
```
surajssd, line 1008: why is this test case removed?

pkg/object/object.go
```diff
@@ -113,17 +144,34 @@ func (e *EnvVariable) validate() error {
...
+	// the following makes sure that exactly one of them exists at a given time
+	if (m.SecretRef != nil) == (m.VolumeRef != nil) {
+		return fmt.Errorf("Exactly one from 'mountRef' or 'secretRef' must be specified for the mountPath: %v", m.MountPath)
+	}
+
+	if m.VolumeRef != nil {
+		// validate volumeRef
+		if err := validateName(*m.VolumeRef); err != nil {
+			return fmt.Errorf("mount %q: invalid name, %v", *m.VolumeRef, err)
+		}
```
surajssd, line 179: validating `secretRef` is missing?

pkg/object/object.go
```diff
@@ -157,11 +205,22 @@ func (c *Container) validate() error {
...
 		// mountPath should not collide, which means you should not do multiple mounts in same place
 		if v, ok := allMounts[mount.MountPath]; ok {
-			return fmt.Errorf("mount %q: mountPath %q: cannot have same mountPath as %q", mount.VolumeRef, mount.MountPath, v)
+			return fmt.Errorf("mount %q: mountPath %q: cannot have same mountPath as %q", *mountType, mount.MountPath, v)
 		}
-		allMounts[mount.MountPath] = mount.VolumeRef
+		allMounts[mount.MountPath] = *mountType
+
+		// validate volumeSubPath
+		// TODO: if there is someway to do it
```
surajssd: these two unnecessary comments have been added, it was already moved to `mount.validate()` ?

pkg/object/object.go
```diff
@@ -157,11 +205,22 @@ func (c *Container) validate() error {
 			return fmt.Errorf("failed to validate mount: %v", err)
 		}
 
+		var mountType *string
```
surajssd: `mountType` is really confusing name, when there is no type associated here but just names.

pkg/transform/util/util.go
concaf: Well, that is why I created this util under `pkg/transform/`, and it is being used in `object.go` and `kubernetes.go`

pkg/object/object.go
concaf: Yeah, because both of them cannot be set at the same time, so setting as a pointer, helps in checking that later, like we have done for other types.

pkg/object/object.go
```diff
@@ -236,9 +295,85 @@ func (v *Volume) validate() error {
...
+	if len(*oSec) == 0 {
+		log.Printf("There are no root level secrets defined, assuming the corresponding secret exists in the cluster: %v", *sRef)
+	} else {
+		secretFound := false
+		for _, secret := range *oSec {
+			if secretDef.SecretName == secret.Name {
+				secretFound = true
+
+				dataKeyFound := false
+				for _, secData := range secret.Data {
```
surajssd: implement a search function on the secret root level object similar to one here in [`VolumeExists`](https://github.com/redhat-developer/opencompose/blob/master/pkg/object/object.go#L92), it makes code more readable!

surajssd: This will also relieve you off of most flags you have used to find if the search was successful or not!

pkg/transform/util/util.go
```diff
@@ -0,0 +1,24 @@
+package util
+
+import (
+	"fmt"
+	"strings"
+)
+
+type SecretDef struct {
+	SecretName string
+	DataKey    string
```
surajssd: I am not sure, this whole thing feels very disconnected to me. Why can't we keep this in the opencompose object? Like we read it as string but we don't need to save it as string, we can split it in the encoding itself.

pkg/object/object.go
concaf: > Where do you validate if the field Base64 is a valid base64 encoded value?

We cannot validate that - https://stackoverflow.com/questions/8571501/how-to-check-whether-the-string-is-base64-encoded-or-not#comment23919648_8571649

> Similarly for field File, it should be a valid path, either absolute or relative depending on what is supported!

We need the path of the OpenCompose file, which we do not have in object.go. This case gets covered in encoding.go, when the absolute secret path gets generated.

pkg/encoding/v1/encoding.go
concaf: That's the logical "and", sort of pseudo code comment :)

pkg/encoding/v1/encoding.go
concaf: make sense, fixed

pkg/encoding/v1/encoding.go
concaf: I have explained it in the comment above. Basically, the relative paths of the secret is getting converted to the absolute path.

pkg/encoding/v1/encoding.go
concaf: make sense, fixed

pkg/encoding/v1/encoding_test.go
concaf: This says that the test should fail if `value` is not provided for the env variable, but this has changed since the value is now a pointer, and the test passed now.
This check is now covered in validations in object.go which checks that exactly one of `value` or `secretRef` has to be provided.

pkg/object/object.go
concaf: That is getting validated in `func (o *OpenCompose) Validate() error`

pkg/object/object.go
concaf: Yep, removed, thanks

pkg/object/object.go
concaf: ack, I have changed this to `mountTypeValue`, and refactored some part of it

pkg/encoding/v1/encoding.go
surajssd: I don't see it though?

pkg/encoding/v1/encoding.go
surajssd: I think the `and` here being programmatic is causing the confusion, I think adding `&&` could remove the confusion.

pkg/encoding/v1/encoding.go
concaf: pushed

surajssd: cool!

pkg/object/object.go
surajssd: We have done for others specifically in `encoding.go` so that we can find what is optional and what is not, is that a case with this particular struct? Since this is populated by us and not the user input.

pkg/object/object.go
surajssd: Anyway the comment in stackoverflow itself explains, and I think that is what we want, we are not checking if the value is base64 encoded of something, we just wanna see if it base64 encoded? We don't care whose!

kadel:
I'm generally OK with this. Once @surajssd gives LGTM it can be merged

pkg/encoding/v1/encoding.go
concaf: fixed

pkg/object/object.go
concaf: Yes, this is being set by us and is being used later on to check if the value is set or not, but do you see a problem with pointers being used here? What do you suggest in place of this and how will that help?

pkg/object/object.go
concaf: There is no way (that I could find) to check if a value is base64 encoded or not with 100% accuracy, we can only check if it _could be_ encoded; I don't really think it serves any purpose to add that check here.

pkg/object/object.go
concaf: Makes sense, implemented this, thanks :)

pkg/transform/util/util.go
concaf: Thinking about this, I agree. I have handled this at the unmarshalling stage, refactored most of this.

concaf (commented):
I have gone ahead and refactored the code to handle the secretRef conversion from string to struct at the unmarshalling time itself.
Also, added docs, fixed tests, added some comments, and minor refactors.

@surajssd @kadel PTAL

pkg/transform/util/util.go
concaf: nvm, removed this anyway

PR #144, opened 2017-05-29
Update reference documentation

cdrage:
This updates the reference documenation in order to have a clearer image
in terms of what's going on with OpenCompose.

We also increase the spacing in-between the titles and rename a few
specification names (ex. PortSpec to Port) for simplicity.

We also switch to example-based learning. Going through each part of the
specification full-example.

### DIFF
--- a/docs/file-reference.md
+++ b/docs/file-reference.md
@@ -1,8 +1,8 @@
-# OpenCompose file reference documentation
+## OpenCompose file reference
 
+_A brief example of the OpenCompose specification:_
 
-OpenCompose file example
-```yml
+```yaml
 version: "0.1-dev"
 
 services:
@@ -21,6 +21,9 @@ services:
     - "/tmp/foobar"
     ports:
     - port: 8080:80
+      type: internal
+      host: domain.tld
+      path: /admin
     mounts:
     - volumeRef: db
       mountPath: /app/store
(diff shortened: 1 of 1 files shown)


cdrage:
ping @pradeepto @surajssd @containscafeine

docs/file-reference.md
```diff
@@ -1,8 +1,8 @@
-# OpenCompose file reference documentation
+## OpenCompose file reference
 
+_A full example of the OpenCompose specification:_
```
kadel: Not sure if we want to show *full* example of OpenCompose specification.
Really soon OpenCompose will have a lot of different features, only small number of complicated apps will use everything. 
A full example might give a wrong impression that OpenCompose is complicated.

docs/file-reference.md
```diff
@@ -37,104 +40,135 @@ volumes:
...
-  args:
-    - "-foo"
-  ports:
-    - <PortSpec>
-  mounts:
-    - <MountSpec>
+### Container
+
+```yaml
+services:
```
kadel, line 138: My original idea was to describe just the given section of the file without distracting it by other "upper"  sections.

Here that would be just what is inside the `<Container>`.

If you think that  was too confusing we can stick with that, I'm ok with that.

docs/file-reference.md
```diff
@@ -169,57 +203,69 @@ This overrides default Command from Docker container image.
...
-### EnvSpec
-```yml
-# <EnvSpec>
-  name: DB_USER
-  value: bob
+### envVariables
+
+```yaml
+...
+env:
```
kadel: this should probably show everything from the top, so it's consistent with rest of the examples.

```yaml
services:
- name: foobar
  ...
  containers:
    ...
    env:
    - name: foo
      value: bar
    ...
  ...
```

docs/file-reference.md
```diff
@@ -169,57 +203,69 @@ This overrides default Command from Docker container image.
...
-```yml
-# <PortSpec>
-  port: 8080:80
-  type: internal
-  host: domain.tld
-  path: /admin
+### Port
 
+```yaml
+services:
```
kadel, line 257: This should show that containers are in an array

``` yaml
services:
- name: foobar
  ...
  containers:
    - ...
      ports:
        - port: 8080:80
          type: internal
          host: domain.tld
          path: /admin
    ...
  ...
```

docs/file-reference.md
```diff
@@ -268,103 +318,125 @@ Path based ingresses specify a path component that can be compared against a URL
...
-  volumeSubPath: foo/bar
-  readOnly: true
+### Mount
+
+```yaml
+...
+services:
+- name: foobar
+  ...
+  containers:
```
kadel, line 333: Same as my comment in ports. It should show that containers are in an array

docs/file-reference.md
```diff
@@ -268,103 +318,125 @@ Path based ingresses specify a path component that can be compared against a URL
...
-| type | required |
+
+| Type | Required |
 |------|----------|
 |string|    yes   |
 
 Name of the EmptyDir volume. 
 
 
 ## volumes
```
kadel: This is top level section. It should have same headline as `services` and `version` and you prepend others with "Section: ".
It should be same "Section: Volumes"

kadel (commented):
Thank you for cleaning up this a bit.
Just a few questions and pointers :innocent:

docs/file-reference.md
cdrage: @kadel A bit confused, I believe that line 138 example (what you comment on here), goes over every possible value of <Container>, having the full example on the top gives users a quick reference / being able to quickly go over and iterate through each value.

Putting "-services" and "-name:foobar" shows WHERE `-container` goes.

docs/file-reference.md
cdrage: Done.

docs/file-reference.md
cdrage: done.

docs/file-reference.md
cdrage: Ah, missed that. Thanks.

docs/file-reference.md
cdrage: Done!

cdrage:
Thanks for the reviews @kadel 

I've gone ahead and updated the PR.

PR #147, opened 2017-05-30
abstract out validate() method for object.Mount

### LINKED ISSUE #142: Abstract `object.Mount` tests in object.go
We need to abstract away tests for `object.Mount` from `func (c *Container) validate() error` to `func (m *Mount) validate() error`, and call from `func (c *Container) validate() error`

parking here for later

concaf:
This abstracts out the validate() methoud for Mount type in
object.go, as has been done for other types in the file.

Also, added relevant tests.

Fixes #142

### DIFF
--- a/pkg/object/object.go
+++ b/pkg/object/object.go
@@ -119,6 +119,24 @@ func (e *EnvVariable) validate() error {
 	return nil
 }
 
+func (m *Mount) validate() error {
+
+	// validate volumeRef
+	if err := validateName(m.VolumeRef); err != nil {
+		return fmt.Errorf("mount %q: invalid name, %v", m.VolumeRef, err)
+	}
+
+	// mountPath should be absolute
+	if !path.IsAbs(m.MountPath) {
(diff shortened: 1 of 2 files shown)


kadel:
Ah :rofl: 
This is similar to what I  suggested to @surajssd while reviewing his volumes PR.
see https://github.com/redhat-developer/opencompose/pull/88#discussion_r112960763

kadel:
I agree with basic idea, but i think that you shouldn't pass `allMounts` as argument to function.
Its required only for checking colliding paths. And that one chack can be done in ` (c *Container) validate() `. as it requires information form container, its kind of container validation.
`(m *Mount) validate() error` should validate only single mount nothing more.

concaf:
@kadel agreed, fixed, PTAL :)

kadel:
This is exactly how I was imagining it ;-) 
What do you thing @surajssd ?

surajssd (approved):
LGTM!

PR #162, opened 2017-06-07
Update examples from tomaskral/nonroot-nginx to nginx

cdrage:
Update the examples to reflect the official nginx package on Docker Hub.

### DIFF
--- a/examples/hello-nginx-external.yaml
+++ b/examples/hello-nginx-external.yaml
@@ -2,7 +2,7 @@ version: 0.1-dev
 services:
 - name: helloworld
   containers:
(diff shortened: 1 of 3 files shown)


kadel (changes requested):
You have to change ports.

nonroot-nginx is listening on 8080 (it can't listen on 80)

But official nginx is using  port 80.


I'm still not a fan of official nginx image :(
But if you think that it is really better to use official, let's fix ports, and we can merge it.

cdrage:
Updated the examples.

I know, but having people use `nginx` would be more beneficial as that's a container that's widely used / people are used to.

kadel (approved):

PR #163, opened 2017-06-08
Add more straightforward README example and general clean-up

cdrage:
Moved the example to the top as a "use case".

Added more widgets.

Cleaned up some other parts of the README (shell autocompletion, etc.)
in order to avoid repetition in README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,55 @@
 # OpenCompose
 
-[![Build Status](https://travis-ci.org/redhat-developer/opencompose.svg?branch=master)](https://travis-ci.org/redhat-developer/opencompose)
+[![Build Status Widget]][Build Status] [![GoDoc Widget]][GoDoc] [![Slack Widget]][Slack]
 
-The goal of OpenCompose is to make it easier for developers to on-board to Kubernetes.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,24 +1,55 @@
...
+Create (or download) `hello-nginx.yaml`
+
+```yaml
+version: 0.1-dev
+services:
+- name: helloworld
+  containers:
+  - image: nginx
+    ports:
+    - port: 8080:80
```
kadel: nginx on docker hub is listenning on port 80.

This should be `80:8080` or just `80`

kadel (changes requested):

README.md
cdrage: Shouldn't it still be `8080:80`?

Ex: ` docker run --name some-nginx -d -p 8080:80 some-content-nginx` ?

EDIT: nvm, I'm an idiot.

cdrage:
@kadel Updated!

README.md
kadel: yeh, It can be bit confusing :-( if you come from docker/docker-compose. OpenCompose has a different order for ports. But there is a good reason for it ;-)

kadel (approved):

